PR #27, opened 2024-01-18
Ldap bind as user

klelifo:
LDAP authentication with a dedicated user for the application itself seems needlessly complicated.
The downside of this approach is that it requires the LDAP admin to create a user for HAWKI on the LDAP side and give it read access to all user entries - which is also a security concern.

With this pull request I propose to use a bind-as-user approach that usually does not need any actions on the LDAP side.
All that is needed is that users are allowed to read their own entry - which is usually allowed by default.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,9 +2,8 @@
 Authentication="LDAP"
 ; LDAP config
 LDAP_HOST="ldaps://..."
-LDAP_BASE_DN="cn=...,ou=...,dc=..."
-LDAP_BIND_PW="..."
-LDAP_SEARCH_DN="ou=...,dc=..."
+LDAP_PATTERN="uid=%username%,ou=people,dc=example,dc=com"
+LDAP_FILTER="(objectClass=*)"
 
 ; Open Ai config
 OPENAI_API_KEY="sk-..."
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .env
 .DS_Store
+.idea
 
 feedback/*/***
--- a/README.md
+++ b/README.md
@@ -51,22 +51,21 @@ To generate answers HAWKI uses the Open AI api. Follow the instructions on https
 
 To get started you need to add a configuration file to the project first. Copy the file ".env.example" from the root directory and rename it to ".env". Replace the example values in it with your own configuration. A detailed description of all values is listed below.
 
-| Value            | Type    | Example                                | Description                                                                                                                                        |
-| ---------------- | ------- | -------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------- |
-| Authentication   | string  | 'LDAP' or 'OIDC'                        | Authentication method: LDAP or OpenID Connect                                                                                                      |
-| LDAP_HOST        | string  | "ldaps://...de"                        | The URL of your LDAP server.                                                                                                                       |
-| LDAP_BIND_PW     | string  | secretpassword                         | Password of the user that is trying to bind to the LDAP Server.                                                                                    |
-| LDAP_BASE_DN     | string  | "cn=...,ou=...,dc=..."                 | Distinguised name that is used to initially bind to your LDAP server.                                                                              |
-| LDAP_SEARCH_DN   | string  | "ou=...,dc=..."                        | Distinguished name that is used for authenticating users.                                                                                          |
-| OIDC_IDP          | string  | "https://...."                         | URL of the Identity provider supporting OpenID Connect.                                                                                            |
-| OIDC_CLIENT_ID    | string  | "..."                                  | Client Id for this application in Identity provider.                                                                                               |
-| OIDC_CLIENT_SECRET | string  | "..."                                 | Secret key for OpenID Connect. 
-| OIDC_LOGOUT_URI | string  | "https://...."                                 | URL to logout from Identity provider                                                                                                                  |
-| OPENAI_API_KEY   | string  | sk-...                                 | Open AI Api key                                                                                                                                    |
-| IMPRINT_LOCATION | string  | https://your-university/imprint        | A link to your imprint. Alternatively you can replace the file index.php under /impressum with your own html/ php of your imprint.                 |
-| PRIVACY_LOCATION | string  | https://your-university/privacy-policy | A link to your privacy policy. Alternatively you can replace the file index.php under /datenschutz with your own html/ php of your privacy policy. |
-| TESTUSER         | boolean | `false`                                | Can be set to `true` for testing purposes. You can then sign in using username `tester` and password `superlangespasswort123`                      |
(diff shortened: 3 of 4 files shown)


nicoroeser:
Hello @klelifo , and thanks for this pull request! Even though I am not a member of the HAWKI project team, my team at Uni Regensburg uses HAWKI, and I found your pull request interesting.

Allow me to give some feedback:
1. Your commit 9e27722397ad2629e4c6e546138ee864f4d35f77 affects a file (directory) which is not generated/used by this project, in this case, `.idea`. I think the commit should not be part of this pull request.

   See the [gitignore documentation](https://git-scm.com/docs/gitignore) for details. Consider using one of the two local gitignore mechanisms, dependent on whether your use of IDEA affects just work on this repository or lots of repositories on your computer.
2. Your approach of including `%username%` looks like a start, but it no longer allows LDAP lookups if, for example, you allow users to log in to HAWKI with their email address and password but your LDAP directory uses something different in the DN (like a *nix username). [`ldap_read`](https://www.php.net/manual/en/function.ldap-read.php) does a search with scope base and thus just having the configurable filter does not solve the problem. A possible solution is to allow the username also in the filter.

   I agree that the [hardcoded filter](https://github.com/HAWK-Digital-Environments/HAWKI/blob/ffd93493ad38ba5eef63df6c9436f3f085175936/login.php#L63) is not nice and should be improved (it certainly is an issue for us at Uni Regensburg). However, for the reason above, I do not think this pull request make it into HAWKI in the current form (but let’s wait for response from the maintainers).

   Some environments might require one specific user for LDAP lookups, others may allow anonymous searching. Some will require lookups, others (like yours) may not need them. (And probably almost all of them will use LDAP bind as the user in order to test password validity.) Due to such variances it may be a good idea to make these things configurable.

jonesvan:
Hello @klelifo,

Firstly, thank you for your contribution and interest in improving HAWKI. Your feedback is invaluable to us as we strive to make HAWKI more accessible and secure for all users. I appreciate the detailed explanation and the time you've taken to propose these changes.

Addressing your concerns:

1. **.idea Directory Commit**: You're correct; the inclusion of IDE-specific directories or files like `.idea` in the repository should generally be avoided unless they contribute to the project's setup or configuration in a meaningful way. We'll ensure to remove this from the pull request and add it to our `.gitignore` file to prevent future occurrences.

2. **LDAP Authentication Approach**: Your suggestion to adopt a bind-as-user approach for LDAP authentication is insightful. This method indeed simplifies the process by eliminating the need for a dedicated application user in the LDAP directory, which, as you pointed out, can be a security concern and an additional administrative burden.

    The flexibility of allowing users to log in with their email addresses while our LDAP directory uses a different attribute in the Distinguished Name (DN) is an important aspect. Your proposal to include the username in the LDAP filter alongside the DN pattern configuration could provide the necessary flexibility for handling different LDAP schemas and authentication requirements.

3. **Configurability**: I concur with your observation that the LDAP configuration requirements can vary significantly across different environments. Some might allow anonymous searches, while others might require a specific user for LDAP lookups. The necessity of making these aspects configurable is clear, and your feedback aligns with our vision to make HAWKI adaptable to various institutional needs.

In response to your feedback and the challenges you've outlined:

- We will remove the unintended commit affecting the `.idea` directory and update our repository's `.gitignore` settings accordingly.
- We plan to enhance the LDAP authentication mechanism to be more flexible and secure, incorporating your suggestions about the bind-as-user approach and the configurable LDAP filter. This will involve updating our documentation to guide administrators through the configuration process, ensuring they can tailor the authentication mechanism to their specific LDAP environment without compromising security.
- We will introduce configuration options for LDAP authentication, including settings for anonymous searches, specific user lookups, and customizable DN patterns and filters. This aims to accommodate the diverse LDAP configurations across institutions, providing a more versatile and user-friendly setup.

Your contribution underscores the importance of community feedback in the ongoing development of HAWKI. We are committed to making these improvements and will prioritize them in our development roadmap. We invite you to continue the discussion and contribute to the implementation of these enhancements if you're interested.

Again, thank you for your valuable input, and we look forward to your continued involvement with HAWKI.

Best regards,

Jonas
HAWKI Team

Here is a simplified idea to approach these requirements:

```php
session_start();

// Function to log messages
function logMessage($message) {
    // Placeholder for logging function
    // Implement logging to a file or a system logger
    error_log($message); // Simple example using PHP error log
}

if (isset($_SESSION['username'])) {
    header("Location: interface.php");
    exit;
}

if (isset($_POST["account"]) && isset($_POST["password"])) {
    function auth($username, $password) {
        $envFilePath = ".env";
        $env = file_exists($envFilePath) ? parse_ini_file($envFilePath) : false;
        
        $host = $env ? $env["LDAP_HOST"] : getenv("LDAP_HOST");
        $pattern = $env ? $env["LDAP_PATTERN"] : getenv("LDAP_PATTERN");
        $filter = $env ? $env["LDAP_FILTER"] : getenv("LDAP_FILTER");

        if (empty($username) || empty($password)) {
            logMessage("Authentication attempt without credentials.");
            echo "Fehler: Keine Anmeldedaten<br>";
            return false;
        }

        $connection = ldap_connect($host);
        if (!$connection) {
            logMessage("Failed to connect to LDAP server: $host");
            echo "Fehler: Verbindung zum LDAP-Server konnte nicht hergestellt werden.<br>";
            return false;
        }
        
        ldap_set_option($connection, LDAP_OPT_PROTOCOL_VERSION, 3);
        // Example for setting LDAP_OPT_REFERRALS option
        ldap_set_option($connection, LDAP_OPT_REFERRALS, 0);

        $sanitizedUsername = ldap_escape($username, "", LDAP_ESCAPE_FILTER);
        $user_dn = str_replace("%username%", $sanitizedUsername, $pattern);
        
        if (!ldap_bind($connection, $user_dn, $password)) {
            logMessage("Authentication failed for user: $user_dn");
            echo "Fehler: Authentifizierung fehlgeschlagen<br>";
            return false;
        }

        $searchFilter = str_replace("%username%", $sanitizedUsername, $filter);
        $result = ldap_search($connection, $user_dn, $searchFilter) ?: false;

        if ($result && ldap_count_entries($connection, $result) > 0) {
            $info = ldap_get_entries($connection, $result);
            $userinfo = $info[0];
            // Set session or perform additional user information processing here
            $_SESSION['username'] = $userinfo["cn"][0]; // Example usage
        } else {
            logMessage("User information fetching failed for user: $user_dn");
            echo "Fehler: Benutzerinformationen konnten nicht abgerufen werden<br>";
            return false;
        }

        ldap_unbind($connection);
        return true;
    }

    if (auth($_POST["account"], $_POST["password"])) {
        header("Location: interface.php");
        exit;
    } else {
        echo "Anmelden fehlgeschlagen";
        http_response_code(401);
        exit;
    }
}

Issue #24, opened 2023-12-18
Double Asterisk deletes content from response


jsubhi:
If ChatGPT surrounds response with double asterisks (**) the content between the asterisks will be deleted from the response.

Prompt to ChatGPT: `Please write "Test: **with asterisk** without asterisk`
Response: `Test: without asterisk`
Expected: Response: `Test: **with asterisk** without asterisk`

(Seen in Implementation at Uni Hildesheim)

msbls:
Same here. For example:
![image](https://github.com/HAWK-Digital-Environments/HAWKI/assets/54026448/09e01859-76a8-459f-a0c9-34814c15f492)

The headline at "2." gets removed as soon as it's written out:
![image](https://github.com/HAWK-Digital-Environments/HAWKI/assets/54026448/c7ad6afe-6e84-4531-afe7-02f1734b251f)

msbls:
You can easily check this behavior if you ask ChatGPT:
"Can you tell me which markup elements you are using in your answers?":
![image](https://github.com/HAWK-Digital-Environments/HAWKI/assets/54026448/a5f03bdd-3ab7-4ef7-9fa7-42f2f9b7ab24)
should include the rendered markup like here:
![image](https://github.com/HAWK-Digital-Environments/HAWKI/assets/54026448/537169ee-32ae-48e0-941c-b367543b5eb3)

msbls:
It looks like GPT 4 uses ** as markup for headlines/bold text.
This line to detect code blocks removes ** output https://github.com/HAWK-Digital-Environments/HAWKI/blob/34c60dd1cf7bcb8de2c47457731c4c8f15a677d6/interface.php#L393 which results in removing headlines/bold text.

HAWK-Digital-Environments:
We have fixed this problem in the latest version of HAWKI. Thanks for the tip!

PR #30, opened 2024-01-31
detect code blocks again

msbls:
It looks like you accidentally removed the code block detection while fixing https://github.com/HAWK-Digital-Environments/HAWKI/issues/24, so I added it again as separate line.


msbls:
Fixed in commit 8038957

PR #40, opened 2024-02-29
Download 3rd party libs into vendor/ directory

### LINKED ISSUE #37: Self-contained installation without use of CDNs should be possible
Thanks for HAWKI; one of our teams at the [Regensburg University](https://www.uni-regensburg.de/en) [Computer Centre](https://www.uni-regensburg.de/computer-centre/home/) is evaluating it and working on a test deployment.

We have found that HAWKI by default makes the clients pull data from Cloudflare and Google CDNs (hosts `fonts.googleapis.com` and `cdnjs.cloudflare.com`). This is not compatible to our privacy standards and is annoying to handle in the context of EU privacy laws (e.g. we need to mention it in the privacy notice and obtain user consent).

I take it for granted that you would welcome pull requests covering these problems. As HAWKI code in its current state is meant to be deployed directly, and does not include a “prepare” or “build” step (see #36 for more on this), I would like to discuss the best way to go forward first so that we do not invest resources on this and risk that the specific implementation suggested by us does not gain approval:

* One option is that some build system and preparatory step is added to HAWKI (→#36) and that the affected files could be pulled from somewhere and included in the local deployment. But users could also manually configure the file sources; they should have a few words of documentation about how to do this.
* In case of some automation: would you prefer that the files are _always_ provided by the web server on which HAWKI is running (my current personal preference), or would you like a _switch_ which selects CDN/own server? (Maybe even allows CDN selection? – e.g. Fira Sans could be taken from Google or Mozilla CDN.)
* If there will be such a switch, what should the defaults be?
* Expected additions to the documentation?
* Further thoughts about this?

jojoob:
... and load it from there instead from external sources.

Fixes #37 

I took the easiest route and just downloaded the currently linked sources, downloaded them into ./vendor/ and load them from there. Not the most elegant way but a quick one.


nicoroeser:
Thanks! I have a somewhat bad feeling about this: if the downloaded files are added to this repo (like in the current pull request), I think that licensing information needs to be added.

If they are not added to the repo but downloaded by users, they will have to add licenses because they are redistributing the files. This calls for automation, or, at least some documentation?

Ariansdf:
Hello @jojoob 
I personally prefer external sources specially during the development to keep the project lighter and the development process more agile. 
However, I manually reviewed your suggestion and added a copy of the libraries to the project. as @nicoroeser has mentioned we needed to also mention the MIT Licenses in the project (see Readme file).
I hope it can solve your problem and thank you for your contribution.

PR #41, opened 2024-03-01
change mouse cursor when over the copy button

msbls:
To make clear that this is a clickable button.
image
-> 
image


Ariansdf:
@msbls thanks for your contribution and mentioning the issue. I think the easier way would be to just add the cursor style directly in the css.
I corrected that in the project and it will be included in the next release.
Thank you

PR #43, opened 2024-03-03
Fix "headers already sent"

ItsNotYou:
`interface.php` sends HTML before trying to establish a session and trying to forward the user to the login page. If the user is not logged in, this results in the following error:

```
PHP Warning:  session_start(): Session cannot be started after headers have already been sent in /var/www/interface.php on line 9
PHP Warning:  Cannot modify header information - headers already sent by (output started at /var/www/interface.php:1) in /var/www/interface.php on line 11
```

This pull request moves the PHP code before any HTML output and fixes this bug.


Ariansdf:
@ItsNotYou The issue is already noted and will be included in the next release. Thank you.

PR #48, opened 2024-03-14
Integrate Microsoft Azure AI

ItsNotYou:
The REST API for Microsoft Azure AI is very similar to the OpenAI API. With minor changes, HAWKI also supports Microsoft Azure AI. The following changes were made:

* Move API URL into configuration
* Send API key in `api-key` header
* Extend README

This pull request breaks backward compatibility, because `OPENAI_API_URL` has to be set via config or environment variable. If required, backward compatibility may be restored with a more complex if else in `stream-api.php:27`.

Compatibility tested with both OpenAI and Microsoft Azure AI.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,8 +6,9 @@ LDAP_BASE_DN="cn=...,ou=...,dc=..."
 LDAP_BIND_PW="..."
 LDAP_SEARCH_DN="ou=...,dc=..."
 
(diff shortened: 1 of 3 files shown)


Ariansdf:
@ItsNotYou thank you for your contribution. I will merge your pull request into the development branch.

PR #71, opened 2024-05-27
Fix: Use references in foreach loop to correctly update feedback data

StenSeegel:
This pull request addresses an issue in the `submit_vote.php` file where the feedback data was not being correctly updated.

Changes:
- Updated the foreach loop to use references when finding the feedback entry by ID. This ensures that any changes made to the `$foundFeedback` variable are reflected in the original `$feedbackData` array.

This change is crucial for the correct functioning of the voting system as it ensures that the updated feedback data is correctly written back to the JSON file.

Please review and merge if everything looks good.


Ariansdf:
Hey @Sten
Thank you for the keen observation. Of course you're right the loop needs to get the references and update the array.
I just updated the development branch and I'll merge your pull request into development.

PR #78, opened 2024-07-15
configurable languages

GaRaOne:
the layout and language files are not yet university-independent.

unfortunately we do not have the time to adapt all languages. therefore we have decided on a subset and have made this configurable in the .env.

this also has the advantage that settings.php no longer contains any magic/redundant knowledge about the supported languages.

since php is not really my everyday language, it would be especially good to check if the following check is correct at this point:
if ( !count( $activeLanguages ) ) {

### DIFF
--- a/private/.env.example
+++ b/private/.env.example
@@ -49,3 +49,4 @@ FAVICON_URI=""
 
 # Default Language, leave blank to use default language of the user's browser.
 DEFAULT_LANGUAGE="de_DE"
+ACTIVE_LANGUAGES="de_DE en_US es_ES fr_FR it_IT"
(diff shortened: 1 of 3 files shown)


Ariansdf:
Hey Steffen,
Thanks for the pull request.
I believe a more "robust" way to do the "if" would be checking it with empty. like:

```    
    // Split the string into an array and remove empty elements
    $activeLanguages = array_filter(explode(' ', $activeLanguagesStr));

    if (empty($activeLanguages)) {
        // Default if no active languages are set
        return ['de_DE', 'en_US', 'es_ES', 'fr_FR', 'it_IT'];
    }
```
 
 and IMHO I think exploding the array with empty spaces can later cause some troubles. when someone for example forgets to put a space there or writes it with comma and they think that the code doesn't work. why didn't you use commas?

PR #85, opened 2024-08-27
Feat: Custom Language Files

fsesterh:
This PR adds a path for custom language files under private/custom/language. Users of HAWKI can place .json language files here and add the language variables they want to change to these custom files. Only those lang variables users want to change need to be present in a custom lang file. I.e. "private/custom/language/de_DE.json". Content of custom lang files is merged with the default language files present under "private/resources/language". If no custom lang file exists the default lang file is used. 

Addes /private/custom to .gitignore to allow users to freely modify custom lang files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 .env
 .DS_Store
-/private/custom
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .env
 .DS_Store
+/private/custom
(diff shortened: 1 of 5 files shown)


sushifrick:
I tested this PR in a local instance and I am really happy with it. The implementation of custom strings is straight forward and patching changed strings during upgrades is not longer necessary. Great work @fsesterh !

Looking forward to see this PR upstream.

PR #88, opened 2024-09-24
added: stream-api-bild.php, new model switch case in interface.php

StenSeegel:
Hi, I have uploaded our Dall-E and GPT-4o1-mini implementation for cross-checking.


Ariansdf:
Hey Sten,
thanks for the PR. I think it's most logical to make a new branch for image generation. so we can keep it parallel to the development.

PR #178, opened 2025-08-29
Improved GPT-5 support by using the response, rather that the completion endpoint

steffenhauf:
This PR adds improved GPT-5 support by using the response endpoint, rather than the completions endpoint.

See e.g. https://platform.openai.com/docs/guides/latest-model for the rational behind this.

**The following is added/modified:**

- add an `OpenAIResponsesProvider` provider which uses the responses endpoint
- add a possibility to retain the response id from openAI as part to the encrypted content of each message. This allows to use the `previous_response` feature, through which GPT-5 models can reuse reasoning results.
- add options to do the aforementioned in a secure way through appropriate configuration settings.

**How I've tested this:**

I've tested this locally againt GPT-5, and GPT-5-mini, both in a streaming and non-streaming setup

### DIFF
--- a/app/Http/Controllers/StreamController.php
+++ b/app/Http/Controllers/StreamController.php
@@ -56,7 +56,6 @@ class StreamController extends Controller
                 'payload.messages.*.role' => 'required|string',
                 'payload.messages.*.content' => 'required|array',
                 'payload.messages.*.content.text' => 'required|string',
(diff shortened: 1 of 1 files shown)


steffenhauf:
What's the plan here. Any interest in this PR. We've since progressed quite a bit on our fork at https://github.com/European-XFEL/HAWKI but would still be interested in merging upstream before the repos diverge too much.

PR #212, opened 2025-10-24
Feat/responses api v2

StenSeegel:


Ariansdf:
Hey Sten,
As we talked before you have pushed the whole project in the PR.
Can you please send another one, only including the OpenAI parts?
Thanks as always

PR #259, opened 2026-01-29
Hotfix web search bug

Ariansdf:
- Added extra condition to fix object reference null for Web search button, when Web Search Active flag is false.
- Documentation update with indication for disabling default web search model

### DIFF
--- a/_changelog/2.3.3.md
+++ b/_changelog/2.3.3.md
@@ -1,0 +1,5 @@
+# v2.3.3
+
+### Bugfix
+
+- Resolved a frontend issue that occurred when no models with web search capability were available.
--- a/_documentation/2-GettingStarted/1-Local Installation.md
+++ b/_documentation/2-GettingStarted/1-Local Installation.md
@@ -178,11 +178,20 @@ then navigate to `config/model_lists/ollama_models.php` and add your model attri
 
 >Before hosting models, make sure your system has the minimum required resources.
 
+>**IMPORTANT**
+> 
+>Currently, HAWKI is configured to use Gemini models’ native web search tool. If you want to disable the Google provider, make sure to add the following environment variables to disable the default web search model:
+>``` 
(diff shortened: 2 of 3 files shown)


public/js/message_functions.js
```diff
@@ -761,7 +761,8 @@ async function regenerateMessage(messageElement, Done = null){
     else{
         inputContainer = messageElement.closest('.thread').querySelector('.input-container');
     }
-    const webSearchActive = inputContainer.querySelector('#websearch-btn').classList.contains('active');
+    const webSearchBtn = inputContainer.querySelector('#websearch-btn') ?? null;
+    const webSearchActive = webSearchBtn ? webSearchBtn.classList.contains('active') : false;
```
Neunerlei, line 765: Minor nitpick: while valid, you could be less verbose here:

```
const webSearchBtn = inputContainer.querySelector('#websearch-btn');
const webSearchActive = webSearchBtn && webSearchBtn.classList.contains('active')
```

public/js/groupchat_functions.js
```diff
@@ -114,7 +114,9 @@ async function onSendMessageToRoom(inputField) {
         const aiKeyRaw = await exportSymmetricKey(aiKey);
         const aiKeyBase64 = arrayBufferToBase64(aiKeyRaw);
 
-        const webSearchActive = inputField.closest('.input-container').querySelector('#websearch-btn').classList.contains('active');
+        const webSearchBtn = inputField.closest('.input-container').querySelector('#websearch-btn') ?? null;
+        const webSearchActive = webSearchBtn ? webSearchBtn.classList.contains('active') : false;
+
```
Neunerlei, line 119: I know why you do this twice, but I am not a fan of duplicating code for basically the same feature. Ok for now

Neunerlei (approved):

PR #281, opened 2026-03-17
Fix: external API crashes on every request (undefined `stream` key)

willirath:
`handleExternalRequest` validation strips `stream` from the payload, but all three provider converters access `$rawPayload['stream']` directly, causing an `ErrorException` -> 500 on every call. This is a regression of the fix in 2.0.1 (be45666), lost again during ae539aa. PR adds the same one-line fix (`$payload['stream'] = false`) as be45666 plus a regression test ensuring that a request to api/req-ai doesn't just get a catch-all 500 response.

This is a draft PR as I don't plan to get involved with HAWKI development but still want to provide a full fix and tests which is awkward in an issue. Feel free to close immediately.

### DIFF
--- a/app/Http/Controllers/StreamController.php
+++ b/app/Http/Controllers/StreamController.php
@@ -53,6 +53,7 @@ class StreamController extends Controller
         }
 
         $payload = $validatedData['payload'];
+        $payload['stream'] = false;
 
         // Handle standard response
         $response = $this->aiService->sendRequest($payload);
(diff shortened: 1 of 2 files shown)


Neunerlei:
Hey! Thanks so much for doing that work even without wanting to officially contribute.
This issue flow around my radar for a  while now, and I now fixed it once and for all. 
As we plan on adding new endpoints that allow streaming, I did not opt for your solution, but instead fixed it where it broke: https://github.com/hawk-digital-environments/HAWKI/commit/e4df8276d7427d0154d5a6ba4939444013b5a7ec 

Nevertheless, thanks for bringing it back to my attention!

Will be fixed in the next release.

PR #287, opened 2026-04-13
test-HKI-126: add debugging config & doc. Add tests for dev cli.

whplh:
- Add remote debugger config in dev
- Updates dev tooling in bin
- Adds tests (jest) for dev tooling in bin
- Running tests: `./bin/env test:cli` or  `cd bin/_env && docker compose -f docker-compose.test.yml run --build tests`
- Format dev tooling code with prettier


Neunerlei:
@whplh :scream: bekommen wir das auch hin ohne das halbe Projekt neu zu formatieren XD?

whplh:
@Neunerlei Done. Merge?

Dockerfile
Neunerlei, line 1: Please do this in `/docker/app/php/php.dev.ini` Your "host.docker.internal" is already there.

vite.config.js
Neunerlei, line 1: Can you please explain why this is required in your situation?

bin/docker-compose.test.yml
Neunerlei, line 1: I like that you are testing in a docker container, but I am not a fan of putting the docker-compose into the "bin" directory. Could you move both docker related files to `bin/_env`, so it is clear that those files are not for testing the project itself?

bin/_env/index.ts
```diff
@@ -1,4 +1,10 @@
+import { Application } from '@/Application.ts';
 import * as process from 'node:process';
-import {Application} from '@/Application.ts';
 
-await (new Application()).run(process.argv);
+export async function main(argv: string[], app = new Application()) {
+  await app.run(argv);
+}
+
+if (import.meta.url === `file://${process.argv[1]}`) {
```
Neunerlei: I assume you do this, so you can use "main" in your test?
But I can not find "main" in any other file in this PR. 
Could you please explain why you did this?

bin/_env/__tests__/env.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+const execAsync = promisify(exec);
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+describe("CLI utils with .env.example as .env", () => {
+    const runEnvScript = async () => {
+        const BIN_PATH = resolve(__dirname, "../../env");
+        return await execAsync(`sh ${BIN_PATH}`, {
+            env: {
+                ...process.env,
+                CI: "true", // Note: There is no reasonable way execute "env" with cli input in tests
```
Neunerlei: Could you please rename it to "NO_INPUT", "TESTING" or "SILENT" or something. Because CI means something different in this context that it normally means.

Neunerlei (changes requested):

vite.config.js
whplh: Without it I get a CORS error for the HMR server in development setups

bin/_env/index.ts
whplh: Ah yes I wanted to  test `main`, but then the test setup and mocking became somewhat messy. `env` loads and sets a lot of required variables
So I switched to running env directly in tests.
The change can be reverted. But I thought it is still not a bad change.

Neunerlei: This is purely philosophical: 

But from a testing standpoint, wouldn't t it make more sense to re-export `Application` here, so the testing code could then do:
```
import {Application} from 'index.ts';
(new Application()).run(['foo', 'bar'])
```
I currently do not see the benefit of exporting the function here? 

As a second point: I would at least put some kind of comment above: `if (import.meta.url === `file://${process.argv[1]}`)` telling people why we are doing this, as it took me a hot second to figure out what you were trying to do (and I wrote the code XD)

vite.config.js
Neunerlei: Okay, I assume it does not hurt.

bin/docker-compose.test.yml
whplh: Alright, I moved it

bin/_env/index.ts
whplh: Ok I reverted the changes since they are not really needed here.

bin/_env/__tests__/env.spec.ts
whplh: -> `TESTING_SKIP_CLI_INPUT`

Dockerfile
whplh: I added a .gitignore file to make overriding the config git friendly

Neunerlei: I mean, sure but in general, I liked what you did here, I would have simply moved into another file :) 

If you want to make it easily adjustable you could put in a `docker/app/env/xdebug.dev.env` file with 
```
XDEBUG_MODE="${XDEBUG_MODE:-debug}"
XDEBUG_START_WITH_REQUEST="${XDEBUG_START_WITH_REQUEST:-yes}"
XDEBUG_CLIENT_HOST="${XDEBUG_CLIENT_HOST:host.docker.internal}"
XDEBUG_CLIENT_PORT="${XDEBUG_CLIENT_PORT:9003}"
XDEBUG_LOG="${XDEBUG_LOG:-/var/www/html/storage/logs/xdebug.log}"
XDEBUG_LOG_LEVEL="${XDEBUG_LOG_LEVEL:-7}"
```

and in: /docker/app/php/php.dev.ini
```
xdebug.mode=${XDEBUG_MODE}
xdebug.start_with_request=${XDEBUG_START_WITH_REQUEST}
xdebug.client_host=${XDEBUG_CLIENT_HOST}
xdebug.client_port=${XDEBUG_CLIENT_PORT}
xdebug.log=${XDEBUG_LOG}
xdebug.log_level=${XDEBUG_LOG_LEVEL}
```

This way you can modify those through using just your .env file.
I think that would be pretty neat. 
Wanna do that?

whplh: ~~seems like envs are not supported in ini files. So now the debug target is created from a template in `docker/app/env/xdebug.dev.env.sh`~~
Sry docs seem to allow it. Will check again.

whplh: [Docs](https://www.php.net/manual/en/configuration.file.php) says defaults are also supported

PR #288, opened 2026-04-21
test: add phpunit tests

whplh:
Adds example tests for external api. 
Run tests: `./bin/env ci:phpunit`


Neunerlei:
This will break so hard agains: https://github.com/hawk-digital-environments/HAWKI/tree/feature/v2.5.0
Please delay this for a while.

whplh:
I rebased the branch to feature/v2.5.0 and updated the new test.
fyi the tests from feature/v2.5.0 do not run successfully in the service container. If they run locally they depend on some local setup/settings or models in the db.

whplh:
(~ sry took me some rounds to cleanup the rebase.)

whplh:
The pr now includes fixes and tests for the `app:token` and `app:removeuser` commands

whplh:
@Neunerlei fyi I cannot login as a testUser on this branch any more.
The login can be handled by handleLogin, but then the code crashes here because the test user is not in the database: https://github.com/hawk-digital-environments/HAWKI/blob/development/app/Http/Controllers/AuthenticationController.php#L103-L105

whplh:
Ok the token creation was not fully fixed. And since the service was mocked in tests it did not show up.
Turns out dependency injection can cause issues with dynamic data. The user in the constructor was never the actual user that was used in the command. 
~ I moved the injection "one level up" to the `AuthFactory`  ( and updated the tests)

PR #289, opened 2026-04-22
fix: replaced outdated models

Ariansdf:
Model lists updated


Neunerlei:
Manually merged with 2.5

PR #299, opened 2026-06-01
bug: Fix input placeholder user message in handshake

### LINKED ISSUE #290: Translation error (DE) for Datakey input field
Thanks for creating such a nice project! I tried it but I ran into two problems with the data key, that seem to be not intended:

- [ ] When a datakey exists and a user logs in, the texts of the prompt still asks for the creation of a datakey, at least in German ("Erstelle hier einen Datakey.", HS_EnterPasskeyMsg).
~- [ ] This is a bigger issue that significantly complicates the workflow: Pasting the Datakey is not possible, so the usage of passkey/password managers is disabled. This leads to the use of short, insecure passwords, which seems to be contradictory to the security enhancements made with the usage of data keys.~

I hope that this is the correct place for putting such an issue - if not, just let me know.

Thank you very much in advance!

whplh:
Fix for #290

### DIFF
--- a/resources/views/partials/gateway/handshake.blade.php
+++ b/resources/views/partials/gateway/handshake.blade.php
@@ -17,7 +17,7 @@
                 <div class="password-input-wrapper">
                     <input
                         class="passkey-input"
-                        placeholder="{{ $translation['Reg_SL5_PH1'] }}"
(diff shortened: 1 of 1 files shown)


Neunerlei:
Done in 2.5

PR #300, opened 2026-06-11
fix: Fix formatting issues for tool call mid stream

### LINKED ISSUE #297: Incomplete Latex formula when using Open AI models
In polymer gels, dynamic light scattering is often applied to reveal inhomogeneites. The result of such studies are apparent diffusion coefficients, but often, people also talk about dynamic correlation lengths. How are the two related?

...
In essence, the apparent diffusion coefficient can be related to the dynamic correlation length via:

\[ D = \frac} \]
		
where:

kBkB​ is the Boltzmann constant,

TT is the absolute temperature,
...

whplh:
Fix for issue from https://github.com/hawk-digital-environments/HAWKI/issues/297
POC: 
[tool-call-mid-stream-breaks-format.html](https://github.com/user-attachments/files/28847274/tool-call-mid-stream-breaks-format.html)

### DIFF
--- a/public/js/ai_chat_functions.js
+++ b/public/js/ai_chat_functions.js
@@ -148,6 +148,7 @@ async function buildRequestObjectForAiConv(msgAttributes, messageElement = null,
     let msg = "";
     let messageObj;
     let metadata;
+    let toolCallInProgress = false;
 
     // Start buildRequestObject processing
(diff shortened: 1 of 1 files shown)


Neunerlei:
This is resolved in v2.5

PR #301, opened 2026-06-25
bug: Fix tool call recursion on empty server tool_call array

### LINKED ISSUE #302: Fix external chat & Fix tool call exceptions for normal conversations for GWDGNonStreaming requests in some models
Bug description:
1) External chat does not work
2) The GWDG Server returns an empty tool_call array for some models which triggers tool call handling on HAWKI side.
This causes exceptions for normal conversations without tool calls via the external chat api (`api/ai-req`)

Actual behaviour:
Empty tool calls are treated as a tool call causing 'Maximum tool calling rounds reached, aborting to prevent potential infinite loop'

Expected behaviour:
Empty tool calls in response should not trigger tool calling handling on HAWKI side.

Affected versions:

    development
    2.4.0

whplh:
fix for https://github.com/hawk-digital-environments/HAWKI/issues/302

### DIFF
--- a/app/Services/AI/Providers/Google/GoogleRequestConverter.php
+++ b/app/Services/AI/Providers/Google/GoogleRequestConverter.php
@@ -54,7 +54,7 @@ readonly class GoogleRequestConverter
             'model' => $modelId,
             'system_instruction' => $systemInstruction,
             'contents' => $formattedMessages,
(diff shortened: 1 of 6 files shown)


Neunerlei:
No longer relevant with 2.5

PR #307, opened 2026-07-14
dev-bug: Fix 'stale' user resolution from singleton. Fix ollama api url. …

whplh:
- SystemContextBootingMiddleware changes the default driver mid-request — it authenticates the Bearer token, then calls setDefaultDriver('sanctum') but Guard singleton returned no user
-  vite app.js -> app.ts
- adjust ollama urls since ollama provider builds chat url from base url https://github.com/laravel/ai/blob/259b92a7b339dbc849c8bc3055042760a31f7efc/tests/Feature/Providers/Ollama/BaseUrlTest.php#L9-L32

### DIFF
--- a/app/Models/Scopes/Traits/UserAwareScopeTrait.php
+++ b/app/Models/Scopes/Traits/UserAwareScopeTrait.php
@@ -8,7 +8,6 @@ namespace App\Models\Scopes\Traits;
 use App\Http\Middleware\SystemContextBootingMiddleware;
 use App\Models\User;
 use App\Services\System\Container\SystemEnvironment;
-use Illuminate\Contracts\Auth\Guard;
 use Illuminate\Http\Request;
 
 trait UserAwareScopeTrait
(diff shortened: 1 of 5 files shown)


app/Models/Scopes/Traits/UserAwareScopeTrait.php
```diff
@@ -22,7 +21,7 @@ trait UserAwareScopeTrait
     public function initializeUserAwareScopeTrait(SystemEnvironment $environment): void
     {
         $this->uast_isRunningInConsole = $environment->runningInConsole();
-        $this->uast_currentUserResolver = static fn(Guard $auth) => $auth->user();
+        $this->uast_currentUserResolver = static fn(Request $request) => $request->user();
```
Neunerlei: Yeah, no.
Sadly this will not work. 
This is exactly how I did it initially, but then noticed, that when the app is running in a CLI, we don't have a request object (or at least none that would carry any information). 

See: `app/Console/Commands/CreateSanctumTokenForUser.php` line 84.
It sets the user on the guard, but the request is not being back-propagated here. 
According to the docs, using the Guard here is the correct solution. 

If there is an issue with my implementation we must fix it in `SystemContextBootingMiddleware. 

@whplh Would you please be so kind to explain what the issue you encountered was? Maybe we can figure out a better way together.

Neunerlei (changes requested):

app/Models/Scopes/Traits/UserAwareScopeTrait.php
whplh: Yes I should have added more insights to the context of the issue and thank you for the insights why its like that.  

Basically the issue is that any api request with a token runs into an exception.

As far as I understand it now the type hint (`Guard $auth ...`) resolves once because its a [Singleton](
https://github.com/illuminate/auth/blob/520581cef948d44a865c595aa425a0420e4e0d8d/AuthServiceProvider.php#L39) and does not pick up the changes in `SystemContextBootingMiddleware` , whereas the auth Factory picks those changes up. Is that a better way?

( btw I do not expect an answer until next week ; ), but wanted to keep it separate from other features )

PR #309, opened 2026-07-20
Reuse legacy file converter interface

whplh:
- Use fileconverter 3.0.0 release https://github.com/hawk-digital-environments/hawki-toolkit-file-converter/
- Reuse and refactor synchronous file converter interface ( Use cache concepts from KreuzbergConverter for service metadata)
- Unify timeout and use nginx (60s) defaults as default timeouts ( but keep GWDG as before at 240 for existing custom setups/installs)

Ref: [HKI 160](https://digital-environments.atlassian.net/browse/HKI-160)

### DIFF
--- a/app/Console/Commands/FileStorageConverterTypesList.php
+++ b/app/Console/Commands/FileStorageConverterTypesList.php
@@ -27,7 +27,26 @@ class FileStorageConverterTypesList extends Command
      * Execute the console command.
      */
     public function handle(FileConverterInterface $converter): void
-    {
+    {   
+
+        $config = app()->get('config');
+
+        $converterType = $config->get('file_converter.default');
+        $converterConfig = $config->get("file_converter.converters.$converterType");
+        if(empty($converterConfig['class'])){
+            $this -> warn("No converter class configured for '{$converterType}'.");
+            return;
+        }
+        $converterClass = $converterConfig["class"];
+
+        if(!$converterClass::isValidConfig($converterConfig)){
+            $this -> warn("Converter config for '{$converterType}'");
+            return;
+        }
+
+        $converter = app()->make($converterClass, ['config' => $converterConfig]);
+        $converter->setConfig($converterConfig);
+
         if (!$converter->isAvailable()) {
             $this->warn('No file converter is currently installed or available.');
             return;
--- a/app/Console/Commands/FileStorageConverterTypesList.php
+++ b/app/Console/Commands/FileStorageConverterTypesList.php
@@ -1,7 +1,10 @@
 <?php
 
+declare(strict_types=1);
+
 namespace App\Console\Commands;
 
+use App\Services\FileConverter\Interfaces\FileConverterExtensionInterface;
 use App\Services\FileConverter\Interfaces\FileConverterInterface;
 use Illuminate\Console\Command;
 use Symfony\Component\Mime\MimeTypes;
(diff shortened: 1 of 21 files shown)


app/Console/Commands/FileStorageConverterTypesList.php
```diff
@@ -27,7 +27,26 @@ class FileStorageConverterTypesList extends Command
      * Execute the console command.
      */
     public function handle(FileConverterInterface $converter): void
-    {
+    {
```
Neunerlei: Why is this here? 
You already have the converter in the method parameter?

app/Services/FileConverter/Handlers/HawkiDocConverter.php
```diff
@@ -3,32 +3,84 @@
...
+    /**
+     * Collects and returns a cached version of the service description endpoint of the 
+     * HAWKI file converter endpoint (`/`
+     * @return array  
+     */
+    protected function getServiceInfo(): array
+    {
+        return $this->cache->remember(
+            key: $this->cacheKey,
+            ttl: (new Carbon($this-> clock->now()))->addDay(),
```
Neunerlei: `$this-> clock` should be `$this->clock`

Neunerlei (changes requested):
Generally good work, but please explain the FileStorageConverterFileTypesList change, I don't get why the code is there?

app/Console/Commands/FileStorageConverterTypesList.php
whplh: Ah yes I was irritated because the command on development  merges the types from `ImagePreProcessingConverter` and I wanted only to have the types from the configured Fileconverter class.  

For the old fileconverter (pdf & doc files) it e.g. returns:
```
application/acrobat
application/ico
application/illustrator
application/msword
application/nappdf
application/pdf
application/photoshop
application/postscript
application/vnd.adobe.illustrator
application/vnd.ms-word
application/vnd.openxmlformats-officedocument.wordprocessingml.document
application/x-msword
application/x-pdf
application/x-photoshop
image/bmp
image/ico
image/icon
image/pdf
image/photoshop
image/psd
image/svg
image/svg+xml
image/tiff
image/vnd.adobe.photoshop
image/vnd.microsoft.icon
image/x-bmp
image/x-eps
image/x-ico
image/x-icon
image/x-ms-bmp
image/x-photoshop
image/x-psd
text/ico
zz-application/zz-winassoc-doc
```

But it was mostly for my understanding so just revert?
Or split like: 
- FileConverterInterface: ...
- FileConverterService: ...

Neunerlei: Ah, I get it.
I think we should keep it, as it makes sense.

What I would do:
1. Create a new Interface: `FileConverterExtensionInterface` which has a `getInnerConverter()` method. 
2. Apply that interface to `ImagePreProcessingConverter` 
3. In `filestorage:converter:types:list` check if `$converter` is of type `FileConverterExtensionInterface` if so, iterate trhough the `getInnerConverter()` until we reached the inner one. 
4. On the way call the `getAllowedMimeTypes` for each step, then do an "array_diff" in the reverse order. Which leaves you with the list of the actual types allowed by each step on the way.

This way you can create a nice output like:

> The file conversion is handled in the following chain,
> which is normally handled in a first come, first serve order (depends on the implementration). 
> 
> **1. \App\Services\FileConverter\Utils\ImagePreProcessingConverter**
> image/jpg
> image/png
> ...
> **2. \App\Services\FileConverter\Handlers\HawkiDocConverter**
> application/acrobat
> application/ico
> application/illustrator
> ...

The benefit of this approach is, that the command is still agnostic to how the converters are instantiated, but it is still able to derive the information it needs through the shared contract.

Neunerlei (changes requested):

app/Services/FileConverter/Handlers/HawkiDocConverter.php
whplh: I ran csfixer and it complained about the @ operator because its globally discouraged. I added an exemption for: https://github.com/hawk-digital-environments/HAWKI/pull/309/changes/06acfb3c72bff8128b56cb9a3f6caab5e78abac8#diff-9b4d9f39ce76da7f7b196c0d4b1b618fc3411bd5548316c31c4f24f2c088e1cfR5
The tests were written by ai. ~ Fixtures and Fake objects look plausible to me. `parseBlocks` there is somewhat debatable, but I added it as is.

PR #312, opened 2026-07-22
fix: set default error message for non-improvable prompts

Neunerlei:
Previously I let the model generate a error message, but because we use a really cheap model it responded crap... 
image
Now it uses a static message instead.


resources/language/chat_de_DE.json
```diff
@@ -161,6 +161,7 @@
                 "cancelEdit": "Bearbeitung abbrechen",
                 "cancelResponse": "KI-Antwort abbrechen",
                 "improvementError": "Verbesserungsvorschläge konnten nicht geladen werden",
+                "improvementModelSaidNoError": "Deine Message ist gut. Keine Verbesserungsvorschläge verfügbar.",
```
whplh: `"Deine Message ...` -> `"Deine Nachricht ...` ?

whplh:
LGTM

PR #313, opened 2026-07-22
fix: adjust position of SystemPromptDialog to ensure correct ordering…

Neunerlei:
Moving the prompt component to the bottom, it will be "higher" in the stack, 
fixing this issue:
image

### DIFF
--- a/resources/js/components/chat/composer/SettingsMenu.svelte
+++ b/resources/js/components/chat/composer/SettingsMenu.svelte
@@ -58,12 +58,6 @@
     }
 </script>
 
-<SystemPromptDialog
-    bind:open={systemPromptOpen}
-    value={composerContext.systemPrompt ?? ''}
-    onChange={(newPrompt) => { composerContext.systemPrompt = newPrompt; }}
-/>
-
 {#snippet settingsBody()}
     <div class="settings-body">
         <div class="system-prompt-section">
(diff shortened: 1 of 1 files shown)


resources/js/components/chat/composer/SettingsMenu.svelte
```diff
@@ -187,6 +181,12 @@
     {/snippet}
 </Breakpoint>
 
+<SystemPromptDialog
+    bind:open={systemPromptOpen}
+    value={composerContext.systemPrompt ?? ''}
+    onChange={(newPrompt) => { composerContext.systemPrompt = newPrompt; }}
+/>
```
whplh, line 188: LOKTM (~Quickfix and no z-Index for now)

whplh (approved):

resources/js/components/chat/composer/SettingsMenu.svelte
Neunerlei: My goal is ALWAYS to avoid z-indexing. 
Every time I touch it it becomes a pain in the backside. Especially in a component ecosystem where you don't know how components are stacked into each other I try to find other solutions.

whplh: Ok.
Also that sounds like a design guidline rule that should be added to somewhere.
Most design systems I know use a z-Index hierarchy and then go from there
 ```
:root {
  --z-base: 0;
......
  --z-popover: 600;
  --z-tooltip: 700;
  --z-toast: 800;

  --z-loading: 900;
  --z-debug: 1000;
}

Neunerlei: Yeah, but that breaks down as soon as you have two popovers or tooltips or must open a modal from a tooltip with the popovers staying open in the background.

PR #314, opened 2026-07-22
fix: focus back into input box after file picker actions

Neunerlei:
Basically, if you upload a file or cancel the upload,
the focus jumps back into the input field.
Fixing:
image

### DIFF
--- a/resources/js/components/chat/composer/FilePicker.svelte
+++ b/resources/js/components/chat/composer/FilePicker.svelte
@@ -30,10 +30,12 @@
             target.value = '';
         }
         isAdding = false;
+        composerContext.focusInput();
     }
 
     function handleCancel() {
         isAdding = false;
+        composerContext.focusInput();
     }
 </script>
 


resources/js/components/chat/composer/FilePicker.svelte
```diff
@@ -30,10 +30,12 @@
             target.value = '';
         }
         isAdding = false;
+        composerContext.focusInput();
```
whplh, line 33: Someone might argue same behaviour should apply when I remove an element/file.
~ During remove it makes most sense to me when the list of files becomes empty. 
~ If it is as a simply code change as here maybe its worth it. Otherwise it can be bookmarked as premature overengineering.
What do you think?

Neunerlei: You mean when one clicks on the file pill? 
Hm, I would have to double check how it behaves when you are using the keyboard. 
Because it would be super annoying if you have to re-"TAB" through all the icons to get to the next one again.

whplh: Ah yes true, the topic is user convenience anyway so fine for me.

Neunerlei: @copilot Can you please create a ticket for this and assign it to @SebM2000?
We should improve the keyboard usability:
I checked and the file pills don't have a "focus-ring". If a file item is "clicked" we should remove it and set the focus back to the input. If the enter/space key is pressed we should focus either the next (or if not available previous) item. If no more files are available focus back into the textarea.

Then create a second ticket for @SebM2000 
Which relates to the first one. 
When the "conflict picker" is shown, a.) the focus ring is cropped of at the top. b.) Currently one has to "TAB" three times for each entry in the list. First: Focus on element, Second: Focus on Online status, Third: Focus on Demand Status. You have TAB through every item in the list to even get to the files or tools. Can we improve the UX here? Not sure what would be a good way to do this, tho.

Neunerlei: See #320 and #321

PR #315, opened 2026-07-22
Fix/rename unloaded legacy chat

Neunerlei:
This pull request refactors how conversation name and slug are managed in the `OldUiMessageHistory` class and updates related logic in the chat sidebar button component. The main goal is to simplify state management by deriving the conversation name and slug directly from the conversation object, reducing duplicated state and potential for inconsistencies.

### State management simplification

* Refactored `OldUiMessageHistory` to derive `conversationName` and `conversationSlug` directly from the `_conversation` object, removing separate state variables for these properties. This reduces redundancy and ensures these properties always reflect the current conversation state. [[1]](diffhunk://#diff-5cfe1f7d88ef05c617d3173e04f1cb022f3d0057d412d9bee26d0b9fa5535abcL11-R16) [[2]](diffhunk://#diff-5cfe1f7d88ef05c617d3173e04f1cb022f3d0057d412d9bee26d0b9fa5535abcL45-L59)

### Component logic updates

* Updated `ChatSidebarButton.svelte` to remove the direct dependency on `oldUiMessageHistory` for conversation name and slug, and instead handle chat renames through the `oldUiBridge.onRenameChat` event. This ensures UI updates are consistent with the new state management approach. [[1]](diffhunk://#diff-dd87aeaeba5cca9070dfdd06b94aab45eb1df8c7c3058a27c180e2f502bc9e03L6) [[2]](diffhunk://#diff-dd87aeaeba5cca9070dfdd06b94aab45eb1df8c7c3058a27c180e2f502bc9e03L39-R43)

Fixes: https://discord.com/channels/1314572179798753363/1526167458438844466/1526170597946822716

### DIFF
--- a/public/js/groupchat_functions.js
+++ b/public/js/groupchat_functions.js
@@ -61,8 +61,11 @@ function initializeGroupChatModule(roomsData) {
     });
     window.oldUiBridge.onRenameChat(async (slug, newName) => {
         const formData = new FormData();
-        if (chatName) formData.append('name', chatName);
-        updateRoomInfo(activeRoom.slug, formData);
+        if (newName) formData.append('name', newName);
+        updateRoomInfo(slug, formData);
+        if (window.oldUiMessageHistory.conversationSlug === slug) {
+            window.oldUiMessageHistory.updateConversation({name: newName});
+        }
     });
     window.oldUiBridge.onOpenRoomControlPanel((slug) => {
         openRoomCP(slug);
--- a/resources/js/oldUi/OldUiMessageHistory.svelte.ts
+++ b/resources/js/oldUi/OldUiMessageHistory.svelte.ts
@@ -8,14 +8,12 @@ export class OldUiMessageHistory {
     private sync = new SyncPipeline<{ [LOAD_CONVERSATION_EVENT]: OldUiConversation }>();
 
     private _type: ComposerContextType = $state('room');
-    private _conversation: OldUiConversation | null = null;
+    private _conversation = $state<OldUiConversation | null>(null);
     private _systemPrompt = $state('');
-    private _conversationName = $state('');
-    private _conversationSlug = $state('');
     private _isInConversation = $state(false);
 
-    public readonly conversationName = $derived.by(() => this._conversationName);
-    public readonly conversationSlug = $derived.by(() => this._conversationSlug);
+    public readonly conversationName = $derived.by(() => this._conversation?.name ?? '');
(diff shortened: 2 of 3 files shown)


resources/js/oldUi/OldUiMessageHistory.svelte.ts
Neunerlei, line 1: Don't know why I implemented it like it was, I now made the conversation itself reactive which will now update all derived values correctly.

I had the issue that the permissioning for rooms were not calculated correctly.

resources/js/snippets/ChatSidebarButton.svelte
Neunerlei, line 1: fix the actual issue by listening on the rename event
This is flaky as it also renames the chat even if the update request to the backend failed. I opted for simplicity here, this can be made better when we have a real data representation in the frontend and don't rely on the old ui for this anymore.

whplh:
fyi in the chat I see: Missing translation: legacy.aiChat.noResponse

whplh:
There is a similar issue in the group chats. ( I did not look to much in the code for the problem)
But I even get an error in the console:
```
Uncaught (in promise) ReferenceError: chatName is not defined
    initializeGroupChatModule groupchat_functions.js:64
    onRenameChat OldUiBridge.svelte.ts:330
    trigger SyncPipeline.ts:51
    triggerRenameChat OldUiBridge.svelte.ts:325
    onNameChange ChatNameMenu.svelte:37
   ```
   
   
Bildschirmfoto vom 2026-07-23 10-31-03

**Comment by Neunerlei** [comment]
> fyi in the chat I see: Missing translation: legacy.aiChat.noResponse

This happened when a request was aborted. Should be fixed by https://github.com/hawk-digital-environments/HAWKI/pull/322

**Comment by Neunerlei** [comment]
@whplh 
Weird stuff... should work now, tho.

File: public/js/groupchat_functions.js [hunk 1]
```diff
@@ -61,8 +61,11 @@ function initializeGroupChatModule(roomsData) {
     });
     window.oldUiBridge.onRenameChat(async (slug, newName) => {
         const formData = new FormData();
-        if (chatName) formData.append('name', chatName);
-        updateRoomInfo(activeRoom.slug, formData);
+        if (newName) formData.append('name', newName);
+        updateRoomInfo(slug, formData);
```
whplh, line 65: ~ `updateRoomInfo` is async but to make an optimistic state update without server feedback is fine for me here

whplh (approved):

PR #324, opened 2026-07-31
Feature/update dependencies

Neunerlei:
Updates all dependencies (backend + frontend)
and does some adjustments + fixes after that.

Ah, and in the Tests I renamed the namespaces from "AI" to "Ai". The directory was already renamed in 2.5.0 and it showed a warning when running composer, so I fixed it, too.


composer.json
```diff
@@ -21,35 +21,36 @@
...
         "laravel-json-api/non-eloquent": "^4.3",
-        "laravel/ai": "*",
-        "laravel/framework": "^13.0",
-        "laravel/reverb": "^1.6.0",
-        "laravel/sanctum": "^4.2.0",
-        "league/flysystem-aws-s3-v3": "^3.30.1",
-        "league/flysystem-sftp-v3": "^3.30.0",
-        "league/flysystem-webdav": "^3.30.0",
-        "logiscape/mcp-sdk-php": "^1.7",
+        "laravel/ai": ">=0.10.1",
```
whplh, line 29: This is an unlimited upper bound. Use Caret version fix ? `^0.10.1`

whplh:
Would you mind adding the translation in this pr?
-> Missing translation: chat.composer.toolMenu.manageTools

composer.json
Neunerlei: That is the composer default for versions < 1.0.0 Can change it, but the next update would change it back.

PR #325, opened 2026-08-03
Extend python mimes for plain text treatment

whplh:
I took a random file from a python github repo and it was `'text/x-script.python' => self::PYTHON,` ~> so add some more python mimes from files found in the wild


Ariansdf (approved):
Looks good!
I just read about the non standard file types. But where do they come from?
Back to the security question, shouldn't we add more restriction on these types?
Just asking...

whplh:
> Looks good! I just read about the non standard file types. But where do they come from? Back to the security question, shouldn't we add more restriction on these types? Just asking...

Ok yes, it makes sense not to add more types than strictly needed. I was having an issue with files being `text/x-script.python` The other ones are rare and not strictly needed. Only if a user needs/wants them, they can be added.

PR #338, opened 2026-08-10
Update OpenAI Models & fix SettingsMenu.svelte not disabling the slid…

max-sp-jlu:
…ers if they are not supported

### DIFF
--- a/_changelog/next.md
+++ b/_changelog/next.md
@@ -2,7 +2,7 @@
 
 ### What's New
 
-[//]: # (- The main new features and changes in this version.)
+- Update `openai_models.php` to include GPT-5.6 Luna, Terra and Sol.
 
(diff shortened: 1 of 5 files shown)


Neunerlei (approved):

Neunerlei:
Hey @max-sp-jlu 
Code wise this looked fine, so I already merged it.
However, when testing I would suggest two UX changes: 
image

1. Below the "Einstellungen" title, add an alert box (we currently don't have a ui component for that as far as I am aware @Ariansdf did you create one in your assistant branch? If not, please create one), with a short text explaining why those inputs don't work. 
2. If the sliders are disabled, I would disable the info icons as well, somehow it feels like a bug to me, that those popovers work even while the slider is disabled.

PR #340, opened 2026-08-11
Adjust the Disabled state in SettingsMenu.svelte

max-sp-jlu:
grafik


Neunerlei:
Nice!
Three nitpicks tho:
1. The icon is not the right size (compare the triangle size with the size of the pen above it)
2. In this case I would drop the headline. 
3. Wording: "vom gewählten Modell" not "von dem"

Thanks :pray:

PR #353, opened 2026-08-18
Feature/prepare component extraction

Neunerlei:
Prepares the extraction of the resources/js/components directory into it's own npm package.

Merge will probably be painful, because I changed a lot of imports. 

@Ariansdf and @SebM2000 Any components that were created and are not yet in the directory should be moved there.

Otherwise, any objections?


resources/js/components/lib/color-scheme/ColorSchemeContext.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+export type ColorScheme = 'light' | 'dark';
+
+/**
+ * Context value published by {@link provideColorScheme}. Exposes the active
+ * colour scheme as a getter (not a plain field) so the value stays reactive
+ * across the context indirection — reading `.colorScheme` inside a
+ * `$derived` or a component template tracks the host's live value rather
+ * than a one-time snapshot taken when the context was created.
+ */
```
whplh, line 16: Ok it took me a moment to understand the benefits of creating a getter api interface. 
Since this pattern will probably be the default now, can this be formulated a bit narrower/simpler/... ?  
I mean the getter does not make the source reactive. And another option would have been to pass the reactive upstream state itself.
So the getter is actually a (nice) api design boundary decision.

~ Maybe: 
```
/**
 * Context value published by {@link provideColorScheme}.
 *
 * The colour scheme is exposed through a getter so consumers only depend on
 * the current value, not on the host application's state/store object.
 * The getter itself does not provide reactivity; the value returned by it
 * must come from the host's reactive state.
 */
```
and for `provideColorScheme`

```
/**
 * Publishes the active colour scheme to the component subtree.
 *
 * The context stores a stable object whose `colorScheme` property forwards
 * reads to the supplied getter. The context itself therefore does not need
 * to be replaced when the host's theme changes.
 *
 * For changes to propagate reactively to consumers, the supplied getter must
 * read from the host application's reactive state.
 *
 * @example
 * ```svelte
 * <script lang="ts">
 *     import {provideColorScheme} from '@hawk-hhg/hawki-svelte-components';
 *     import {useStore} from '$lib/app/hooks/useStore.svelte.js';
 *
 *     const themeStore = useStore('theme');
 *     provideColorScheme(() => themeStore.theme);
 * </script>
 * ```
 */
```

Neunerlei: That would be a documentation change only?

whplh (approved):

PR #360, opened 2026-08-26
Add `AuthChecker` middleware for route authentication and session val…

max-sp-jlu:
…idation

- Introduce `AuthChecker` to enforce authentication on guarded routes with periodic and focus-based session refreshes.
- Integrate `AuthChecker` into `RouteRegistrar` for middleware assignment based on route meta.
- Extend `StoreExtension` and `DataStore` to support synchronous initialization via an optional `init` method.
- Update `KeychainStore` to initialize keychain handle during app assembly.
- Add route middleware handling to `RouteRegistrar` options.


resources/views/modules/chat.blade.php
```diff
@@ -58,6 +58,12 @@
                     </div>
 
                     <h1 id="start-title">{{ __("StartBanner") }}</h1>
+                    <!-- TODO: For Testing only -->
```
Neunerlei: I would suggest to remove that?

resources/js/kernel/stores/types.ts
```diff
@@ -18,5 +24,7 @@ import type {HawkiApp} from '$lib/kernel/HawkiApp.js';
 export interface DataStore {
     readonly name: string;
 
+    init?(app: HawkiApp): void;
```
Neunerlei: Just a question about naming stuff.
`init` is called when the app is `ready`, meaning when the `ready` event on the bootstrapper is triggered. Would it make sense to name it ready for consistency or keep it as is? @whplh

resources/js/components/ui/routing/logistics/RouteRegistrar.ts
```diff
@@ -269,6 +269,21 @@ export interface RegisteredRouteGroupOptions extends RouteGroupOptions {
     children: (registrar: RouteRegistrar) => void;
 }
 
+/** Options for a {@link RouteRegistrar}; inherited by every nested registrar a {@link RouteRegistrar.group} creates. */
+export interface RouteRegistrarOptions {
```
Neunerlei: This is the "Auth" global middleware in a trenchcoat again.

Neunerlei (changes requested):
Please review and merge/rebase: https://github.com/hawk-digital-environments/HAWKI/pull/370

PR #362, opened 2026-08-26
Svelte chat: reasoning timelines, web-search activity, inline threads and generation stats.

max-sp-jlu:
## Overview
This PR adds reasoning summaries, web-search activity, inline message threads, and optional generation statistics to the Svelte chat frontend. It also reorganizes model settings, improves pasting and conversation scrolling, groups chat history by recency, and introduces shared page components and mobile navigation behavior. Backend streaming and message persistence changes support the new UI.
## Added features
 - TODO 

<details>
  <summary>Click here to view implementation details</summary>

**Reasoning summaries and web-search activity**

- Added `MessageReasoning.svelte`, a collapsible timeline above assistant responses. It separates titled reasoning summaries into steps and displays web searches, opened pages, and searches within pages alongside them.
- Search steps show their query and linked source chips with favicons and URL previews. Sources are deduplicated by domain for display, with the first six visible and additional sources available through a nested expansion. The timeline trigger also shows the number of search actions.
- The panel opens by default during reasoning, highlights the active step, and remembers the user's expanded/collapsed preference in local storage. Timeline and expansion animations respect reduced-motion preferences.
- `OpenAiAdapter` requests automatic reasoning summaries and web-search sources for text-generating agents using models flagged as reasoning-capable. `ExtendedOpenAiGateway` inserts separators between completed summary parts so adjacent sections do not run together.
- `StreamController` emits structured web-search status events with action type, query, and deduplicated source URLs. `ChatTransport` combines these events with reasoning text deltas and updates the message while the response streams.
- Reasoning is saved inside the encrypted message content alongside the answer and citations, then restored by `ChatStore` when reopening a conversation. Loading also accepts reasoning stored as a single string. The UI displays provider-supplied summaries when available and uses translated thinking/generating labels instead of raw stream status keys.

**Optional generation statistics**

- Added the `statsForNerds` experiment and `MessageStats.svelte`. When enabled, assistant responses with recorded metrics show input tokens, output tokens, tokens per second, time to first text token, and total request duration.
- Completion packets now include provider token usage, with corresponding frontend stream types. The client records timing during generation and saves the resulting statistics inside the encrypted message content, so they remain available after reloading.
- Token counts remain unavailable until the provider reports them and display as `n/a` when missing. Throughput uses output tokens divided by the full request duration, accounting for the reasoning phase included in output usage. Timing collection happens independently of the display toggle.

**Inline message threads**

- Added `messageThreads.ts` to group the existing flat message list into parent messages and replies using the legacy `W.DDD` IDs. Pending user messages and streaming assistant messages carry a temporary thread index until persisted IDs are available.
- `ChatMessage` now renders replies beneath their parent in an expandable section with a reply count, a reply action, and an indicator while the composer targets that thread. Threads open when composing or receiving pending/streaming replies; an empty thread closes when thread mode is exited.
- Thread creation is limited to persisted parent messages. Replies whose parent cannot be found remain visible as top-level messages.
- `ChatConversation` renders the grouped messages and excludes thread replies from the scroll behavior that moves a newly sent top-level message to the top of the viewport.

**Model settings and reusable menu controls**

- Rebuilt `SettingsMenu` using the shared `DropdownMenu`, which handles desktop menus and mobile bottom sheets. Split system prompts, reasoning effort, sampling controls, and reset headings into separate components.
- Added a reasoning-effort submenu that lists only the levels advertised by the selected model, ranging from `none` through `max` where supported. Models without adjustable effort hide the control. Request parameters omit an unsupported selection so switching models does not send an incompatible `reasoning_effort` value.
- Kept the Creative, Balanced, and Precise sampling presets and moved the temperature/Top P sliders behind an Advanced settings expander. Models without sampling support hide the presets and show disabled sliders with an explanation.
- Preserved the system-prompt preview, editing dialog, and reset behavior, plus the reset for model parameters. Tab and Shift+Tab cycle through the settings panel's buttons, menu items, tabs, and sliders.
- Added `DropdownMenuSub` with hover and keyboard interaction, portaled content, and a full-width panel below the trigger on small screens. Radio items now support checkmark indicators and keeping the menu open after selection. `SingleSelect` also exposes a bindable `open` state.

**Pasting, rendering, scrolling, and export**

- Pasting plain text longer than 1,000 characters attaches it as a localized `.txt` file when attachments are enabled and the file is accepted. Otherwise, the text stays inline. Ctrl/Cmd+Shift+V explicitly keeps plain text inline, while clipboard files continue through the existing attachment handling.
- Added stable client render keys for pending and streaming messages, retaining them when server IDs arrive. This avoids remounting message components at persistence time.
- Opening a conversation now keeps the view pinned to the bottom as asynchronous Markdown, formulas, and other content change its height. Scrolling away from the bottom or sending a new top-level message releases that pin.
- Corrected the Markdown renderer's `final` flag so only completed content is marked final. Removed stray whitespace introduced by `TextLink` and tooltip markup around inline links.
- Reduced PDF export text from 16px to 14px and removed the explicit `html2canvas` scale setting.

**Conversation history and new-chat behavior**

- Grouped sidebar history into Today, Yesterday, Last 7 days, Last 30 days, and Older, omitting empty groups and preserving the existing newest-first order. Buckets use local calendar dates without daylight-saving shifts affecting the day count; missing or invalid dates fall into Older.
- Added the conversation relationship to `AiConvMsg::$touches`, so message changes update the parent conversation timestamp used for history ordering and grouping.
- Added `ChatStore.requestNewChat()` and the `onNewChatRequested` event. Both the sidebar action and logo use this path, close mobile navigation, and refocus the composer even when the new-chat page is already open.

**Shared page layout and mobile navigation**

- Added reusable `Page` and `PageHeaderBar` components and adopted them in `ChatIndex`, `ChatConversation`, and `ChatHeader`. They support default or custom headers, custom chat bodies, loading spacers, visible or screen-reader headings, and print layouts.
- Shared header styling keeps controls before page content in reading and tab order, reserves space for the mobile navigation toggle, and uses a locally scoped stacking context for the blurred backdrop.
- The mobile navigation button now toggles both open and closed states with matching accessibility labels. `MobileNavCollapse` closes navigation on outside pointer presses while excluding the opening press, navigation controls, scrollbars, and open floating menus.
- Added configurable fades beneath the floating mobile toggle, with shorter or disabled variants for pages that need them. Chat content reserves space so the toggle does not cover the initial messages or welcome screen.
- Widened the composer dock and added a blurred, tinted backdrop across it. Adjusted mobile spacing and removed redundant z-index declarations from the file-drop overlay and chat header controls.

**Development support, shared styling, and translations**

- Added `createHmrSafeContext()`, which uses stable `Symbol.for()` keys across Vite hot updates. Migrated the app, router scope, sidebar, menu list, toast, and composer contexts to prevent providers and consumers losing their shared context after module re-evaluation. Missing composer providers still produce a specific error.
- Added base styles for buttons, disabled buttons, labels, and horizontal rules, plus a matching `u-label` utility.
- Moved the existing `smob` dependency into `devDependencies` and updated the lockfile without changing its version.
- Added English and German translations for history groups, reasoning and search actions, thread controls, pasted-text filenames, reasoning-effort settings, and generation statistics, including the experiment's title and description.
</details>


app/Services/Ai/LaravelAi/Drivers/OpenAiExtended/ExtendedOpenAiGateway.php
```diff
@@ -42,6 +43,13 @@ class ExtendedOpenAiGateway extends OpenAiGateway
      */
     private array $lastData = [];
 
+    /**
+     * Set when a reasoning summary part has finished so the next {@see ReasoningDelta}
+     * can be separated from the previous part. OpenAI streams each summary part
+     * (typically "**Title**\n\nBody") without any delimiter between parts.
+     */
+    private bool $reasoningPartEnded = false;
```
whplh: This looks pretty hacky. Why not add&use:
https://github.com/laravel/ai/blob/0.x/src/Streaming/Events/ReasoningEnd.php

Pls research how to handle that.

app/Services/Ai/Providers/Adapters/Implementations/OpenAiAdapter.php
```diff
@@ -68,6 +71,31 @@ public function getModels(AiProviderProxy $provider): Collection
...
+     */
+    public function getAdditionalDriverOptions(Agent $agent, AgentRequestContext $context): array
+    {
+        if ($agent instanceof AbstractTextGeneratingAgent && $context->model->flags->hasStrengthReasoning()) {
+            return [
+                'reasoning' => [
+                    'summary' => 'auto',
+                ],
+                // Expose the sources a native web search found, so the UI can show them alongside the reasoning.
+                'include' => ['web_search_call.action.sources'],
```
whplh: This gates sources behind the reasoning capabilities. Check if that can be kept separate

whplh (changes requested):

app/Services/Ai/LaravelAi/Drivers/OpenAiExtended/ExtendedOpenAiGateway.php
whplh: Hm ok, the last commit is slightly going towards a better direction, but still way off.
What I meant was to research a protocol on how to implement a thinking protocol with the events that exist:
https://github.com/laravel/ai/tree/0.x/src/Streaming/Events

This is more like what I meant conceptually:
https://github.com/hawk-digital-environments/HAWKI/compare/feature/svelte-qol...feature/svelte-qol-rev-1
( if you use the code do the TODO and test it thoroughly, which also includes the non streaming path, which should or should not disply thinking steps?)

app/Http/Controllers/StreamController.php
whplh, line 1: Also add entries to `_changelog/next.md`. Boil it down to the eesential and newly added features (and the context, e.g for which providers to expect which behaviour change).
The internals section change should help the next dev to understand the core internal changes from this PR. The main goal there is to transport ideas and hint and make another dev curious and aware of code changes concerning the features added here

whplh (changes requested):

whplh:
LGTM now :+1: . Only the TODO in the PR description begs for some attention

whplh (approved):

PR #364, opened 2026-08-27
fix: handle s3 files

whplh:
See [next.md](https://github.com/hawk-digital-environments/HAWKI/pull/364/changes#diff-6fa7afc6cca33365d3f12162438af5bb11cab45afe9320cfe66a4b4e97edca55) changes for summary/overview

### DIFF
--- a/app/Console/Commands/CheckStorageConnection.php
+++ b/app/Console/Commands/CheckStorageConnection.php
@@ -1,332 +1,262 @@
 <?php
 
+declare(strict_types=1);
+
 namespace App\Console\Commands;
 
 use Illuminate\Console\Command;
-use Illuminate\Support\Facades\Storage;
-use League\Flysystem\WebDAV\WebDAVAdapter;
-use League\Flysystem\PhpseclibV3\SftpAdapter;
-use League\Flysystem\PhpseclibV3\SftpConnectionProvider;
-use League\Flysystem\AwsS3V3\AwsS3V3Adapter;
-use Aws\S3\S3Client;
-use Throwable;
-
-
+use Illuminate\Contracts\Filesystem\Factory as FilesystemFactory;
+use Illuminate\Contracts\Filesystem\Filesystem;
 
 class CheckStorageConnection extends Command
 {
+    /**
+     * Name of the temporary test file that is written to and removed from every checked disk.
+     */
+    private const TEST_FILE_NAME = '.storage_connection_test.txt';
+
     /**
      * The name and signature of the console command.
      *
      * @var string
      */
-    protected $signature = 'check:storage {--filesystem=s3 : The filesystem to test (s3, local, public, localstorage, nextcloud, sftp)}';
+    protected $signature = 'check:storage {--filesystem= : Comma-separated list of disks to check explicitly (default: the disks the storage roles resolve to)}';
 
     /**
      * The console command description.
      *
      * @var string
      */
-    protected $description = 'Check Storage Connection for specified filesystem';
+    protected $description = 'Check the storage connection for the disks the app actually uses (list / write / read / delete test)';
 
     /**
      * Execute the console command.
      */
-    public function handle()
+    public function handle(FilesystemFactory $storage): int
     {
-        $filesystem = $this->option('filesystem');
+        /** @var array<string, array<string, mixed>> $configuredDisks */
+        $configuredDisks = (array) config('filesystems.disks', []);
 
-        $this->info("Testing {$filesystem} storage connection...");
+        $checks = $this->resolveChecksToRun(array_keys($configuredDisks));
+        if ($checks === null) {
+            return self::FAILURE;
+        }
 
-        $result = match($filesystem) {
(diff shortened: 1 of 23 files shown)


app/Console/Commands/CheckStorageConnection.php
Neunerlei, line 1: Am I weird, and I know you have not written this file @whplh but, isn't it weird? 
Why are we setting up the instances of the backends ourself? 

`\Illuminate\Filesystem\FilesystemManager::disk` has a method which can get the instance of the actual configured filesystem. 

From that point forward all tests should have the exact same body, as we expect the flysystem filesystem to work the same in all cases?

If you can not be bothered to make this more consistent, please create a new ticket.

Neunerlei (approved):
Looks fine to me, please check the one file comment I had.

app/Console/Commands/CheckStorageConnection.php
whplh: Yes all good points. 
Turns out there are 3 storage configurations which should be checked for a running system:
```
    'default' => env('FILESYSTEM_DISK', 'local'),
    'file_storage' => env('STORAGE_DISK', 'local_file_storage'),
    'avatar_storage' => env('AVATAR_STORAGE', 'public'),
```

The default is not used in HAWKI, but since it is the frameworks default most packages (like laravel/ai-sdk) have a path to it.  
~ I extended the docs and added a warning whenever it is used from a fallback path.

app/Services/Storage/Filesystem/DefaultDiskWarningFilesystemManager.php
```diff
@@ -0,0 +1,62 @@
...
+ * `filesystems.avatar_storage` roles. Code that resolves a disk without an explicit name
+ * (e.g. a bare `Storage::put(...)` or `app('filesystem')->disk()`) is almost certainly a
+ * bug or an uninformed third-party package, so it logs a warning once per request while
+ * still returning the disk (warn-and-continue: the default remains a valid framework fallback).
+ */
+class DefaultDiskWarningFilesystemManager extends FilesystemManager
+{
+    private bool $defaultDiskWarningEmitted = false;
+
+    public function __construct(Application $app, private readonly LoggerInterface $logger)
```
Neunerlei: Using this, we lose potentially already instantiated disks and "customCreators" here.

Could you please use the $app->extend() with the Decorator trait (https://github.com/hawk-digital-environments/HAWKI/blob/development/app/Utils/DecoratorTrait.php)

app/Providers/StorageServiceProvider.php
```diff
@@ -27,6 +28,9 @@ class StorageServiceProvider extends ServiceProvider
 {
     public function register(): void
     {
+        // Replace the framework FilesystemManager with our own which wanrs then implicit default-disk fallback is used
+        $this->app->singleton('filesystem', DefaultDiskWarningFilesystemManager::class);
```
Neunerlei: See the comment in DefaultDiskWarning...Manager

tests/Concerns/CreatesFilesystemMocks.php
```diff
@@ -0,0 +1,37 @@
...
+
+declare(strict_types=1);
+
+namespace Tests\Concerns;
+
+use Illuminate\Contracts\Filesystem\Filesystem;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\MockObject\Stub;
+
+trait CreatesFilesystemMocks
```
Neunerlei: When naming things use ...Trait and ...Interface, so not "CreatesSomething" but "SomethingCreatorTrait", not "Applier" but "StuffApplyingInterface". 

In this case FilesystemMockingTrait or similar. 
Thanks!

Neunerlei (changes requested):
Generally looks good :)
Just three style advices.

app/Services/Storage/Filesystem/DefaultDiskWarningFilesystemManager.php
whplh: I wrapped it in  `DefaultDiskWarningFilesystemManager::decorate`

Another option is to extend the DecoratorTrait to handle optional properties like:

`createDecoratedOf(object $parent, array $extraProperties = []); `

The interface looks nice e.g. :
```
return DefaultDiskWarningFilesystemManager::createDecoratedOf($manager, [
    'logger' => $app->get(LoggerInterface::class),
]);
```

But I'm not sure if its a good idea to handle properties in the DecoratorTrait - wanna go there?

Neunerlei (approved):
Thanks for adjusting! Well done :)

app/Providers/StorageServiceProvider.php
Neunerlei: [DecoratorTrait.withprops.txt](https://github.com/user-attachments/files/31693669/DecoratorTrait.withprops.txt)

PhpStan extensions live at: app/Services/PhpStan

Thanks for that great suggestion :heart:

PR #367, opened 2026-08-28
Add reusable header components. Add mobile nav collapse

whplh:


resources/js/app/components/sidebar/MobileNavCollapse.svelte
```diff
@@ -0,0 +1,56 @@
...
+            // outside the nav's DOM, but a press there belongs to the layer —
+            // switching modules or picking a menu action keeps the nav open.
+            // Every bits-ui layer marks its content with data-state="open".
+            if (path.some((el) => el.dataset.state === 'open')) {
+                return;
+            }
+            sidebar.navOpen = false;
+        }
+
+        document.addEventListener('pointerdown', handlePointerDown);
```
max-sp-jlu: Here you could also use the svelte:document directive, this handles the registration and cleanup for you.

resources/js/app/components/sidebar/MobileNavCollapse.svelte
```diff
@@ -0,0 +1,56 @@
...
+
+    $effect(() => {
+        if (!sidebar.mobile || !sidebar.navOpen) return;
+
+        // Ensure an activation delay: the very press that opened the overlay
+        // must never immediately dismiss it.
+        let activated = false;
+        const timer = setTimeout(() => (activated = true), 0);
+
+        function handlePointerDown(event: PointerEvent): void {
```
max-sp-jlu: Not a fan of having the listener function defined in an effect block.

resources/js/plugins/core/modules/chat/components/ChatHeader.svelte
```diff
@@ -30,7 +30,6 @@
 </script>
 
 <header {...restProps} class={["u-print-hidden", className]}>
-    <h1 class="u-sr-only">{conversation.name}</h1>
```
max-sp-jlu, line 33: Why is this removed?

resources/js/components/ui/page/Page.svelte
```diff
@@ -0,0 +1,204 @@
...
+            </div>
+        {/if}
+    </div>
+    {#if hasBar}
+        <header class="page-header" class:default-bar={!header}>
+            {#if header}
+                {@render header()}
+            {:else}
+                <div class="page-header-bar">
+                    <h2 class="page-title">{title}</h2>
```
max-sp-jlu: nit: Page Title sollte ein h1 sein.

max-sp-jlu (changes requested):

resources/js/components/ui/sidebar/SidebarContent.svelte
```diff
@@ -29,16 +31,15 @@
     id="main-content"
     class={["content", className]}
     tabindex="-1"
-    inert={sidebar.mobile && sidebar.navOpen}
```
max-sp-jlu, line 32: this could leed to accidental clicks, inside the content, when the sidebar is open

resources/js/components/ui/page/Page.svelte
```diff
@@ -0,0 +1,204 @@
...
+            {@render body()}
+        {:else}
+            <div class="page-scroll" class:reserved={!hasBar}>
+                {#if children}
+                    {@render children()}
+                {/if}
+            </div>
+        {/if}
+    </div>
+    {#if hasBar}
```
max-sp-jlu: why is the header behind the main content in the semantics.

resources/js/components/ui/page/Page.svelte
```diff
@@ -0,0 +1,204 @@
...
+        <header class="page-header" class:default-bar={!header}>
```
max-sp-jlu: This also seems like a duplication of the ChatHeader, extract it into a component, this should be extracted into a common component.

max-sp-jlu (changes requested):

resources/js/app/components/sidebar/MobileNavCollapse.svelte
whplh: ok

resources/js/components/ui/page/Page.svelte
whplh: Thats the price when using DOM stacking without z-Index shuffling. Consecutive components  are layered on top of previous ones

resources/js/plugins/core/modules/chat/components/ChatHeader.svelte
whplh: ~Ah yes :+1:  I broke the layout on the assistant branch and this class became visible. So I deleted it to deal with it later ; ) ~ readded it.

resources/js/plugins/core/stores/ChatStore.svelte.ts
```diff
@@ -37,6 +37,8 @@ export class ChatStore implements DataStore {
     public error = $state<string | null>(null);
     /** Conversation slugs with an AI request still running in this tab. */
     public generatingSlugs = $state<string[]>([]);
+    /** Reactive helper to refocus the composer on new chat requests. */
+    public newChatToggle = $state(0);
```
max-sp-jlu: this seems more like a event then a state, maybe Martin's eventsystem could be usefull here

max-sp-jlu (approved):
otherwise, LGTM

resources/js/components/ui/page/Page.svelte
max-sp-jlu: But this creates even more problems with the Screenreader Semantics, because if that, the title would be below the content, in the semantics.

whplh: Good argument. Ok that will only work with a z-index here.

